PR #6, opened 2019-04-23
solution for colored blocks by @laverka

valeriitovstyk:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+div.red {
```
gestap: Можна просто написати `.red`

PR #20, opened 2019-04-23
Fs on apr19 valeraonisko

valeraonisko:


src/style.css
```diff
@@ -0,0 +1,19 @@
+
+.red {
+  background-color: red;
+  height: 100px;
```
YuriiHoliuk: do not repeat styles
will be better to create one general class for block and add height style to it

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+.blue {
+  background-color: blue;
+}
+
+body {
+  margin: 0;
+}
+
+.same {
```
YuriiHoliuk: `same` is not the best name but ok

PR #31, opened 2019-04-24
Roman_On

ronyshchenko:


src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.top {
+  height: 100px;
+  margin: auto;
```
gestap: What this `margin` do?

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  background: red;
+}
+
+.center {
+  height: 100px;
```
gestap: You duplicate css rule. Can we do something with this?

src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
```
gestap: This is bad practice.
Can we do something with this?

PR #34, opened 2019-04-25
Anna Kondratiuk

annakondr:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.r {
```
gestap: Class name with one letter, it's a very bad idea.

PR #46, opened 2019-05-09
Fs on apr19 stas t

s0lm1r:


src/style.css
```diff
@@ -0,0 +1,18 @@
...
+  background: red;
+}
+
+.green {
+  height: 100px;
+  background: green;
+}
+
+.blue {
+  height: 100px;
```
gestap: Try to write style without repeat it

PR #53, opened 2019-05-10
Colored blocks fe_on_jan19_sasha

saniasem:
https://saniasem.github.io/layout_colored-blocks/

https://saniasem.github.io/layout_colored-blocks//report/html_report/


src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
vkryvytskyy: it's default value for all div's

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.red {
```

PR #148, opened 2019-07-22
Kseniia`s task 

KseniiaMazan:
some styled blocks added

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<link rel="stylesheet" type="text/css" href="style.css"> 
+	<title></title>
+</head>
+<body>
+<div id = "red"></div>
+<div id = "green"></div>
+<div id = "blue"></div>
+</body>
+</html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,18 @@
+body {
+	margin: 0px;
+}
+
+div {
+	height: 100px;
+	width: auto;
+}
+
+#red {
+	background-color: red;
+}
+#green {
+	background-color: green;
+}
+#blue {
+	background-color: blue;
+}


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+	margin: 0px;
```
YuriiHoliuk: do not need to specify units with 0, stylelint should reject it🤔

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+div {
+	height: 100px;
+	width: auto;
```
YuriiHoliuk: do not neccessary

src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+	margin: 0px;
+}
+
+div {
```
YuriiHoliuk: do nott add styles to tags, it is too common. Imagine you need to add other div to page, with different styles

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+	height: 100px;
+	width: auto;
+}
+
+#red {
```
YuriiHoliuk: do not style id, use only classes in CSS(ok to use tags for reset like body margin)

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+	background-color: red;
+}
```
YuriiHoliuk: should be empty lines between style blocks

src/index.html
```diff
@@ -0,0 +1,14 @@
+
```
```suggestion
```

src/index.html
```diff
@@ -0,0 +1,14 @@
...
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<link rel="stylesheet" href="style.css">
+	<title>Colored blocks</title>
+</head>
+<body>
+<div class="block red"></div>
```
```suggestion
  <div class="block red"></div>
```

src/index.html
```diff
@@ -0,0 +1,14 @@
...
+<div class="block green"></div>
```
```suggestion
  <div class="block green"></div>
```

src/index.html
```diff
@@ -0,0 +1,14 @@
...
+<div class="block blue"></div>
```
```suggestion
  <div class="block blue"></div>
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.block {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
```

PR #149, opened 2019-07-22
Add three block of colors red, green and blue

amelin0:
Add three block of colors red, green and blue


src/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<head>
+	<meta charset="UTF-8">
+	<link rel="stylesheet" type="text/css" href="style.css"> 
+	<title>Homework2</title>
+</head>
+<body>
+<div id = "red"></div>
```
YuriiHoliuk: do not need too add spaces around `=` in HTML

src/style.css
```diff
@@ -0,0 +1,17 @@
+body {
+	margin: 0px;
```
YuriiHoliuk: do not to specify units with 0 value

src/style.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+#red {
```
YuriiHoliuk: do not use id for styling, use classes instead

src/style.css
```diff
@@ -0,0 +1,17 @@
...
+	background-color: red;
+}
```
YuriiHoliuk: should be empty line between blocks

src/style.css
```diff
@@ -0,0 +1,17 @@
...
+#green {
+	background-color: green;
+}
+#blue {
+	background-color: blue;
+} 
+
+div {
```
YuriiHoliuk: do not style tags, add class to each block instead. div is too common

YuriiHoliuk (changes requested):

src/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<link rel="stylesheet" type="text/css" href="style.css">
+	<title>Homework2</title>
+</head>
+<body>
+<div class="red"></div>
```
```suggestion
  <div class="red"></div>
```

src/index.html
```diff
@@ -0,0 +1,13 @@
...
+<div class="green"></div>
```
```suggestion
  <div class="green"></div>
```

src/index.html
```diff
@@ -0,0 +1,13 @@
...
+<div class="blue"></div>
```
```suggestion
  <div class="blue"></div>
```

src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+div.red {
```
gestap: no need to write `div`
use only class name

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+  background-color: red;
+}
+
+div.green {
```

src/style.css
YuriiHoliuk: you fix Ostap comment, but div(same as other tags) is too general for styling, will be better to add common class

YuriiHoliuk (changes requested):

YuriiHoliuk (approved):

PR #150, opened 2019-07-22
HW-2 Colored blocks

kosbielozorov:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: 100%;
```
YuriiHoliuk: it is default div style, do not need to specify

YuriiHoliuk (approved):

KosBelozyorov:
add fix style.css

PR #151, opened 2019-07-22
colored blocks solution

mozbie:


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: div has 100% width by default

PR #152, opened 2019-07-22
sol

Taisiiapn:


src/style.css
```diff
@@ -0,0 +1,24 @@
+.first {
+  background: red;
+  height: 100px;
+  width: 100%;
+  margin: 0;
```
YuriiHoliuk: width: 100%;
  margin: 0; - defaults for div, do not need to specify these styles

src/style.css
```diff
@@ -0,0 +1,24 @@
+.first {
+  background: red;
+  height: 100px;
```

Taisiiapn: Thanks, I fixed that

src/style.css
```diff
@@ -0,0 +1,20 @@
+.row {
+  height: 100px;
+  width: 100%;
```
gestap: it's a default value, no need to write it

src/style.css
```diff
@@ -0,0 +1,19 @@
+.row {
+  height: 100px;
+}
+
+.first {
+  background: red;
```
gestap: if you add the only color, better to use `background-color`

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.second {
+  background: green;
+}
+
+.third {
+  background: blue;
+}
+
+body {
```
gestap: move common style to the top of the file

Taisiiapn: fixed, thanks

PR #153, opened 2019-07-22
Mate academy colored_blocks

VladGritsenko:


src/index.html
```diff
@@ -0,0 +1,17 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <title>Document</title>
+</head>
+<body>
+  <div class="conteiner">
```
YuriiHoliuk: typo here, should be `container`

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  padding: 0;
+}
+
+.conteiner {
+  display: grid;
+  grid-template-rows: repeat(3, 100px);
+  grid-template-columns: 100%;
+}
+
+.block:nth-child(1) {
```
YuriiHoliuk: it is ok for now, but will be better to use specific classes(block-red `class="block block-red" if you also need general class`) with separate class names you can reuse styles, but sometimes it is ok to use  nth-child

PR #154, opened 2019-07-22
Vika Tsykun solution

ViktoriiaTsykun:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: ok, but div has 100% by default

PR #155, opened 2019-07-22
PR-layout-colored-blocks-fs-jul19-artem-kr

sci-kid:


src/style.css
```diff
@@ -0,0 +1,23 @@
+
+#red-block {
```
YuriiHoliuk: do not style id's, use classes instead

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+  width: 100%;
```
YuriiHoliuk: div has 100% width by default, do not need to specify it

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+.blue-block {
+  background-color: blue;
+}
+
+body {
+  padding: 0;
+  margin: 0;
+}
+
+section {
```
YuriiHoliuk: as I wrote before "use classes instead", "move common styles to common **class**"

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+
+.green-block {
+  background-color: green;
+}
+
+.blue-block {
+  background-color: blue;
+}
+
+body {
```
gestap: better to write common style on the top of the document.
move `body` to the top.

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  padding: 0;
```
gestap: body doesn't have a default `padding`, no need to write it

PR #156, opened 2019-07-22
'Solution'

Volodsher:


src/style.css
```diff
@@ -0,0 +1,22 @@
+#first {
```
YuriiHoliuk: do not use id for styles, also do not use tags, only classes

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: div already has default style `width: 100%`, do not need to specify it

Volodsher:
Thanks for checking, I've fixed mistakes (I believe so).

src/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>3 Colored blocks</title>
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body class="forBody">
```
YuriiHoliuk: use only lowercase letters for css classes, use `-` or `_` to separate words

forBody - not a good name, will be better to name `container` etc.

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  height: 100px;
+  background-color: green;
+}
+
+.third {
+  height: 100px;
+  background-color: blue;
+}
+
+.forBody {
```
YuriiHoliuk: you should add these styles directly to body

src/style.css
```diff
@@ -0,0 +1,19 @@
+.first {
+  height: 100px;
```

Volodsher: I've put extra tags in <head> and moved common styles to common class. Thank you!

PR #158, opened 2019-07-22
Solution

ivansafiulin:


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0px;
+}
+
+[class*="item-"] {
+  height: 100px;
+  width: 100%;
```
gestap: it's a default value for `div`, no need to write it.

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.item-1 {
```
gestap: `item-1` bad name for class

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+
+.item-2 {
+  background: green;
+}
+
+.item-3 {
+  background: blue;
+}
+
+
```
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.item-2 {
+  background: green;
+}
+
+.item-3 {
+  background: blue;
+}
+
```
```suggestion
```

src/style.css
gestap: you need only one empty line on the end of the document

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0px;
```
gestap: if you write `0`, no need to write `px, rem ...`, it`s a mistake for browser.
```suggestion
  margin: 0;
```

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+[class*="item-"] {
```
gestap: it's good but I recommend to use class. It will be easier to read and debug code.

src/style.css
gestap: remove `width`, `div` already have it.

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
+}
+
+.red.block {
```
gestap: No need to write like this. Better will be `.red`

src/index.html
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <title>Document</title>
+</head>
+<body>
+  <div class="red block"></div>
```
```suggestion
  <div class="block-red block"></div>
```

ivansafiulin:
.

PR #159, opened 2019-07-23
Lesson 2 - added div to html and styles to css

Vyacheslav-P:


src/index.html
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>colored blocks</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+<body>
+  <div class = "red"></div>
```
YuriiHoliuk: do not need to add spaces around `=` in HTML

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```
YuriiHoliuk: do not add styles to div, it is too general, use classes for styling(add general class to each block)

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.general {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.general {
+  height: 100px;
+  width: 100%;
```
gestap: it's a default value for `div`, no need to write it.

Vyacheslav-P: replaced from ".general" to ".block"

YuriiHoliuk: Ostap means `div` always have 100% width, you should delete this style rule

PR #160, opened 2019-07-23
Fs jul19 konstantin b

ghost:
layout_colored-blocks solution by K. Burnazaki


src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  width: 100%;
```
YuriiHoliuk: it is default for div, do not need to specify

YuriiHoliuk (approved):

PR #161, opened 2019-07-23
add three color blocks

youvovk:


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+
+  margin: 0;
+}
+
+.block__red {
+
```
YuriiHoliuk: do not need to add empty line at the start of style block

src/index.html
```diff
@@ -0,0 +1,15 @@
...
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <title>Colored-Blocks</title>
+</head>
+<body>
+  <div class="block__red"></div>
+  <div class="block__green"></div>
+  <div class="block__blue"></div>
```
YuriiHoliuk: you try to use BEM, it is great, but color should be modifier and should be written `class="block block--red"`

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  height: 100px;
```

src/style.css
youvovk: but when I check input "npm run lint" it say me add emtpy line at the start of style block. and it say error when I don't that

youvovk: for example: src/style.css
  6:1  ×  Expected empty line before rule   rule-empty-line-before
 12:1  ×  Expected empty line before rule   rule-empty-line-before
 19:1  ×  Expected empty line before rule   rule-empty-line-before
 26:1  ×  Expected empty line before rule   rule-empty-line-before
 31:1  ×  Expected empty line before rule   rule-empty-line-before

youvovk: delete line

src/index.html
youvovk: thank for help, it's really more useful then I did

src/style.css
youvovk: done

PR #162, opened 2019-07-23
1_lesson_1

ArnyCorp:


src/index.html
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <title>Colored blocks</title>
+</head>
+<body>
+  <div class="red" alt="red"></div>
```
YuriiHoliuk: you need to specify alt attribute only on img tags

src/style.css
```diff
@@ -0,0 +1,17 @@
+body {
+  padding: 0;
+  margin: 0;
+}
+div {
```
YuriiHoliuk: add common class instead, div is too general for styles

src/style.css
```diff
@@ -0,0 +1,17 @@
...
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: it default div style, do not necessary to write it

src/style.css
```diff
@@ -0,0 +1,17 @@
...
+}
+.green {
+  background: green;
+}
```
YuriiHoliuk: should be empty lines between blocks

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  padding: 0;
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.green {
```

ArnyCorp:
Ok, understand!

вт, 23 июля 2019 г. в 16:37, gestap <notifications@github.com>:

PR #163, opened 2019-07-23
fs_jul19 Alex Andrutskii solution

aleks-and:


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.block-1,
+.block-2,
+.block-3 {
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: it is default div style, do not need to write it

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.block-1,
+.block-2,
+.block-3 {
```
YuriiHoliuk: ok for 3 classes, but will be better to add general block class in html

YuriiHoliuk (approved):

PR #164, opened 2019-07-23
My solution to the homework "Colored blocks"

VladyslavSak:


src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  width: 100%;
```
YuriiHoliuk: `width: 100%;` - default div style, do not need to specify it

PR #165, opened 2019-07-23
HomeWork2

ibezkorovayniy:


src/style.css
```diff
@@ -0,0 +1,21 @@
+.first {
+  background-color: red;
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: it default style

src/style.css
```diff
@@ -0,0 +1,21 @@
+.first {
+  background-color: red;
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.second {
+  background-color: green;
+}
+
+.third {
+  background-color: blue;
+}
+
+body {
```
gestap: better to write common style on the top of the document.
move `body` to the top.

src/style.css
```diff
@@ -0,0 +1,19 @@
+.height {
```
gestap: it's a bad name for a class

PR #166, opened 2019-07-23
Fs jul19 dmytro m

Deman250:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
gestap: better to add a common class for common style

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  height: 100px;
+}
+
+.block1 {
```
gestap: `block1` bad name for class

src/index.html
```diff
@@ -0,0 +1,15 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Document</title>
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
+    <div class="block-red block"></div>
```
gestap: Better to write like this.
```suggestion
    <div class="block block-red"></div>
```
At the begin, common style then specifical

PR #167, opened 2019-07-23
lesson-1 colored blocks solution

alexDemche:
Oleksii Demchenko lesson-1 colored blocks solution

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,18 @@
+* {
+  margin: 0;
+}
+
+.block-1 {
+  height: 100px;
+  background-color: red;
+}
+
+.block-2 {
+  height: 100px;
+  background-color: green;
+}
+
+.block-3 {
+  height: 100px;
+  background-color: blue;
+}
--- a/src/index.html
+++ b/src/index.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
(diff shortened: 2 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,18 @@
+* {
+  margin: 0;
+}
+
+.block-1 {
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+  background-color: red;
+}
+
+.block-2 {
```
gestap: `block-2` bad name for a class

src/style.css
```diff
@@ -0,0 +1,18 @@
+* {
```
gestap: you only need to change default `margin` on `body`, no need to add this to all elements

src/style.css
alexDemche: Thank yo for review, i corrected mistake.

src/style.css
alexDemche: Thank yo for review, i corrected mistakes

src/style.css
alexDemche: Thank yo for review, i corrected mistakes

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.redBg {
```
YuriiHoliuk: do not use camelCase in CSS, separate words by `-` or `_`

Also, red-bg is ok class for this task, but not for the future

PR #168, opened 2019-07-23
Colored blocks: solution

PastushenkoVik:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.color-block {
+  width: 100vw;
```
YuriiHoliuk: block has 100% width by default

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+}
+
+.color-block:nth-of-type(1) {
```
YuriiHoliuk: instead of using nth-of-type will be better to add specific classes

YuriiHoliuk (approved):

PR #169, opened 2019-07-23
Home_Work CSS

Voronok16:


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0.01px;
```
gestap: what is this? Do not write fractional values

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+  background-color: red;
+  height: 100px;
```

PR #170, opened 2019-07-23
solution Ihor Sobko (task layout_colored-blocks)

igaryok:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.conteiner {
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk: it is default style, do not need to add

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.c1 {
```
YuriiHoliuk: will be better to add more descriptive classes

src/style.css
igaryok: fixed

src/style.css
igaryok: fixed

PR #171, opened 2019-07-23
colored_blocks_solution

eugeneSahach:


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+#red {
+  background-color: red;
+  height: 100px;
```
gestap: you duplicated this rule.
Create a common class for common style.

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+#green {
```

yevhen-sahach:
I've changed all thing that you suggested, please check it out.

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+[class*="block"] {
```
gestap: better to create class instead of this structure `[class*="block"]`.
It would be easier to read and debug it

yevhen-sahach:
done

ost-ptk (approved):

PR #173, opened 2019-07-23
HW-2

Drop777:


src/index.html
```diff
@@ -0,0 +1,19 @@
...
+<head>
+  <meta charset="utf-8">
+  <title>HTML</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+
+<body>
+
+  <div class="red"></div>
+
```
gestap: remove unnecessary empty line

src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+body div {
```
gestap: create a common class for common style. 
Do not style tag, use class for this. Except `body`.

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  min-height: 100px;
+}
+
+.red {
```

PR #176, opened 2019-07-24
Fs on jul19 artem

ArtemMarkevych1:
Lesson - 2 Colored Blocks


src/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
```
gestap: use 2 whitespaces instead of 4

src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  width: 100%;
```
gestap: no need to write this style, `div` already have it.

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  height: 100px;
+}
+
+.red {
```

src/index.html
ArtemMarkevych1: Done: 
image

src/style.css
ArtemMarkevych1: done:
image

src/style.css
ArtemMarkevych1: Ok, got you.

PR #177, opened 2019-07-24
'Solution'

Serhii-P:


src/style.css
```diff
@@ -0,0 +1,21 @@
+html,
```
```suggestion
```

gestap: `html` doesn't have a default margin.

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  margin: 0;
+}
+
+.colors {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
+  background: red;
```
gestap: better to write `background-color` if you add only color.

src/style.css
```diff
@@ -0,0 +1,21 @@
+html,
+body {
+  margin: 0;
+}
+
+.colors {
+  height: 100px;
+  width: 100%;
```
gestap: `div` already have `width: 100%`, no need to write it.

PR #180, opened 2019-07-24
Colored

Vladyslav223:


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
+  margin: 0;
+}
+
+body div {
```
gestap: create a common class for common style. Do not style tag

src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
```
gestap: no need to add all elements `margin: 0`
add it to body

readme.md
```diff
@@ -17,5 +17,5 @@ You can find some sample screenshots in the `backstop_data/bitmaps_reference` fo
 
 **Result**
 
-[UPDATE DEMO LINK](https://<your_account>.github.io/<repo_name>/) <br>
-[UPDATE TEST REPORT LINK](https://<your_account>.github.io/<repo_name>/report/html_report/)
+-[DEMO LINK](https://vladyslav223.github.io/layout_colored-blocks/) <br>
+-[TEST REPORT LINK](https://vladyslav223.github.io/layout_colored-blocks/report/html_report/)
```
irenhh: download the reference image so that the test works

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.coloured {
+  width: 100%;
```
irenhh: here you have width 100% by default, no need to add it

PR #181, opened 2019-07-24
Fs jul19 koretskyi

koretskyi:


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.top {
+  height: 100px;
```
gestap: you duplicate this rule. 
create a common class for common style

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.line {
```
danheim: This is a block element, not a line, the name should be appropriate.

src/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Document</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+<body>
+  <div class="top line"></div>
+  <div class="middle line"></div>
+  <div class="bottom line"></div>
```
danheim: Main classes should be written first.
```suggestion
  <div class="block bottom"></div>
```

PR #182, opened 2019-07-25
New Solution

ivansafiulin:


ivansafiulin:
commit 7700853 contains information about last changes. In commit 
9befe87 applied the rule and in commit  9bfd89 rolled it back. So look at 7700853 commit.

PR #183, opened 2019-07-26
Fs on jul19 shklyar

shklyarigor2015:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
irenhh: do not style div tag, use general class instead

src/style.css
```diff
@@ -0,0 +1,24 @@
+
+body {
+  margin: 0;
+}
+
+.block {
+
```
irenhh: do not leave empty spaces here

PR #184, opened 2019-07-29
Solution

ChernijAnna:


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  width: 100%;
+  margin-left: 0;
+  margin-top: 0;
```
danheim: You can write this

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  width: 100%;
```
danheim: it is 100% by default

src/style.css
danheim: In this situation you can write `margin: 0;`

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  margin-left: 0;
+  margin-top: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
```
danheim: Div has 100% width by default.

PR #197, opened 2019-08-28
Fs aug19 kirill

anisimovv:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+div {
```
gestap: don't style tag, use class for it.
exclusion `body`

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+  padding: 0;
```
gestap: body don't have a default padding, no need for this rule

src/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="ru">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <title>Document</title>
```
gestap: update title

PR #198, opened 2019-08-28
Solution

Lyudmila1711:


src/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title>Document</title>
+	<link rel="stylesheet" type="text/css" href="style.css">
+</head>
+<body>
+<div class="red"></div>
```
gestap: don't forget about whitespaces
```suggestion
  <div class="red"></div>
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```
gestap: don't style tag, use class for it

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  display: block;
```
gestap: `div` block element.
this is a default value for `div`, no need to write it.
https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.first {
+  background-color: red;
+  height: 100px;
+
```
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.first {
+  background-color: red;
+  height: 100px;
```
Try to not repeat yourself.

PR #200, opened 2019-08-28
Fs aug19 sasha g

Shurkodr:
Aleksandr Glazunov


src/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
```
gestap: fixed whitespaces
use 2 instead of 4

src/index.html
```diff
@@ -0,0 +1,13 @@
...
+    <title>Colored blocks</title>
+    <link href="style.css" rel="stylesheet">
+</head>
+<body>
+<div class="block red"></div>
```
```suggestion
  <div class="block red"></div>
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
gestap: it's a default value, no need to write it.

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.red {
```
Also, `block`

PR #207, opened 2019-08-28
Fs aug19 kostia s

kostya-flern:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.red {
```

ReadmeRepo.md
```diff
@@ -0,0 +1,2 @@
+[Demo link](https://kostya-flern.github.io/layout_colored-blocks/)
```
gestap: You already have readme file in repo, no need to create new. You need to update an existing file

PR #208, opened 2019-08-28
Fs on aug19 dima m

Dmitriymush:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin-top: 0;
+}
+
+.blockheight {
+  height: 100px;
+}
+
+.color1 {
```
gestap: In this task ok, but for the future, `color1, color2, color3 ...` bad name for the classes.

PR #209, opened 2019-08-28
Solution

frucelestine:


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
```
Try to not repeat yourself.

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.red {
```

PR #211, opened 2019-08-28
Solution

CrazyNoodl:


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
```
gestap: better to use `body` in this case

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
```

PR #212, opened 2019-08-28
Fs on apr19 liubomyr

lubremets:


src/style.css
```diff
@@ -0,0 +1,19 @@
+
+body {
+  margin: 0;
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+  height: 100px;
```
Try to not repeat yourself.

PR #213, opened 2019-08-29
Fs aug19 kolya

chepa17:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
gestap: `div` has already width:100%, it's default value.
No need to write it.

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.itema {
```
gestap: In this task ok, but for the future, `itema, itemb, itemc ...` bad name for the classes.

PR #264, opened 2019-12-16
Serebrenykov: Added 3 colored blocks

Alex-8-8:
https://alex-8-8.github.io/layout_colored-blocks/
https://alex-8-8.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
gestap: створи клас, не стилізуй `div`

PR #286, opened 2019-12-16
Develop

Annasmile3:
- [DEMO LINK](https://Annasmile3.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Annasmile3.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  padding: 0;
```
danheim: у боди нет паддингов по-умолчанию

PR #293, opened 2019-12-27
solution

Al145:
- [DEMO LINK](https://Al145.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Al145.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+div {
+  height: 100px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+div {
```
YuriiHoliuk: Will be better to create .box class, but ok for this task

PR #294, opened 2019-12-27
three blocks were added

Valeriy-Sorochynskyi:
- [DEMO LINK](https://Valeriy-Sorochynskyi.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Valeriy-Sorochynskyi.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.green {
+  background-color: green;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
YuriiHoliuk: Will be better to create .box class, but ok for this task

PR #295, opened 2020-02-24
solution

romarios1987:
- [DEMO LINK](https://romarios1987.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://romarios1987.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
+}
+
+.block_red {
+  background-color: red;
+}
+
+.block_green {
+  background-color: green;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
vkryvytskyy: Try to avoid selector *. Yes in these tasks it doesn't influence performance, but in bigger projects, it can load browser

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: red;
+}
+
+.block_green {
+  background-color: green;
+}
+
+.block_blue {
+  background-color: blue;
+}
```

romarios1987:
Исправил

PR #296, opened 2020-02-24
solution

hheight:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  width: 100%;
+  height: 100px;
+}
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```

src/style.css
```diff
@@ -2,19 +2,21 @@ body {
   margin: 0;
 }
 
-div {
+.block-red,
+.block-green,
+.block-blue {
```

PR #297, opened 2020-02-24
solution

crash2bash:
- [DEMO LINK](https://crash2bash.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://crash2bash.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,23 @@
+body {
+  margin: 0;
+  padding: 0;
```

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.first-block,
+.second-block,
+.third-block {
```
earthhuman41: will be nice if you create common class for all blocks like .block

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+  height: 100px;
+}
+
+.first-block {
```
earthhuman41: block--red - better name then first-block

8hokage (changes requested):

src/style.css
crash2bash: Agree :)
All fixed

8hokage (approved):

PR #299, opened 2020-02-24
initial commit

coastin:


src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
vkryvytskyy: Follow the indentation rules. Add an additional space of 2 spaces for nested tags.

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```

PR #300, opened 2020-02-24
added 3 blocks

EugeneOd:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.red {
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+  padding: 0;
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.red {
+  height: 100px;
+  background-color: red;
+}
+
+.green {
```

src/style.css
```diff
@@ -1,19 +1,19 @@
...
   margin: 0;
-  padding: 0;
 }
 
-.red {
+.container {
   height: 100px;
+}
+
+.red {
```
earthhuman41: container--red will be better name

PR #301, opened 2020-02-24
second hw

Wincherz:
- [DEMO LINK](https://Wincherz.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Wincherz.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+  
+.block {
+  height: 100px;
+}
+
+.block1 {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+  
+.block {
+  height: 100px;
+}
+
+.block1 {
```
earthhuman41: block--red will be better name than block1

PR #302, opened 2020-02-24
solution colored_blocks

yurez91:


vkryvytskyy:
Добавь ссылки в PR

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+#block2 {
+  background-color: green;
+  height: 100px;
+}
+
+#block3 {
+  background-color: blue;
+  height: 100px;
+}
```

PR #304, opened 2020-02-24
finish task

K1ngp1nDev:


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+#red {
```
earthhuman41: class will be better

earthhuman41: when you will use BEM methodology you will always use classes, so get used to it

K1ngp1nDev:
- [DEMO LINK](https://K1ngp1nDev.github.io/layout_human-rights/) <br>
- [TEST REPORT LINK](https://K1ngp1nDev.github.io/layout_human-rights/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
```
earthhuman41: i think block--red will be better

K1ngp1nDev:
Done

earthhuman41:
nothing changed since my last review

earthhuman41:
Why you ignore this comment 
Screen Shot 2020-02-24 at 3 34 19 PM

K1ngp1nDev:
I did not notice div in style.css

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.div {
```

K1ngp1nDev: what? .div - class div, not tag

earthhuman41: oh, it's bad idea for call it .div. The name of the element class should make a sense. better block

K1ngp1nDev:
mb now?

PR #305, opened 2020-02-24
Blocks were created and styled

Maildes:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.case {
```
earthhuman41: better .block or .box

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+  width: 100%;
```

PR #307, opened 2020-02-24
solution

BorisovecViktor:


earthhuman41:
Add links to description of Pull request

BorisovecViktor:
у меня стоят ссылки 
    - [DEMO LINK](https://BorisovecViktor.github.io/layout_colored-blocks/)
    - [TEST REPORT LINK](https://BorisovecViktor.github.io/layout_colored-blocks/report/html_report/)

BorisovecViktor:
> _No description provided._

BorisovecViktor:
разобрался и пофиксил

src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  margin: 0;
+  padding: 0;
```
earthhuman41: you don't need to reset margin and padding in all elements

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+section {
+  height: 100px;
+  width: 100%;
+}
+
+.first {
```
earthhuman41: first-block will be better name

src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+section {
```

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.first-block {
+  height: 100px;
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+  padding: 0;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.block-size {
+  height: 100px;
+  width: 100%;
```

PR #308, opened 2020-02-24
solution

Vladimir-Beliy:
- [DEMO LINK](https://Vladimir-Beliy.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Vladimir-Beliy.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  width: 100%;
+  height: 100px;
+}
+
+.red {
+  background: red;
+}
+
+.green {
+  background: green;
+}
+
+.blue {
+  background: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  width: 100%;
+  height: 100px;
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.blocks {
+  height: 100px;
+}
+
+.color1 {
```
earthhuman41: i think block--red will be better then color1

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.blocks {
+  height: 100px;
+}
+
+.block1 {
```
earthhuman41: block--red is better name for it

Vladimir-Beliy: fixed

PR #309, opened 2020-02-24
Added 3 colored cubes to index.html and style.css

Aleksandr-Tyagun:


vkryvytskyy:
Добавь ссылки в описание PR

Aleksandr-Tyagun:
> Добавь ссылки в описание PR

Добавил ссылки в описание PR

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.green {
+  height: 100px;
+  background-color: green;
+}
+
+.blue {
+  height: 100px;
+  background-color: blue;
+}
```

Aleksandr-Tyagun: I changed the names of the class elements to "box-1 ..." and created a new class for common characteristics.

vkryvytskyy: @Aleksandr-Tyagun good job

PR #310, opened 2020-02-24
color blocks were added

KatarinaPetrenko:
[DEMO LINK](https://KatarinaPetrenko.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://KatarinaPetrenko.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
+
+.block1 {
+  background-color: red;
+}
+
+.block2 {
+  background-color: green;
+}
+
+.block3 {
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
```

KatarinaPetrenko: Solved.
thanks

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  background-color: green;
+  height: 100px;
+  width: 100%;
+}
+
+.block3 {
+  background-color: blue;
+  height: 100px;
+  width: 100%;
+}
```

KatarinaPetrenko: solved

PR #311, opened 2020-02-24
I did the 2nd task

Qoa11a:
https://Qoa11a.github.io/layout_colored-blocks/
https://Qoa11a.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
+  width: 100%;
+}
+
+.green {
+  background-color: green;
+  height: 100px;
+  width: 100%;
+}
+
+.blue {
+  background-color: blue;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  width: 100%;
+}
+
+.green {
```

src/style.css
```diff
@@ -2,20 +2,18 @@ body {
   margin: 0;
 }
 
-.red {
-  background-color: red;
+div {
```

PR #312, opened 2020-02-24
colors added

Tata2222:
- [DEMO LINK](https://<Tata2222>.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://<Tata2222>.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
Не работают ссылки

Tata2222:
исправлено. ссылки теперь работают

src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```

Tata2222: names of classes changed

src/style.css
```diff
@@ -8,14 +8,14 @@ div {
   width: 100%;
 }
 
-.red {
+.red-box {
```

src/style.css
vkryvytskyy: Это тоже надо исправить

src/style.css
Tata2222: fixed

src/style.css
vkryvytskyy: fix this

PR #313, opened 2020-02-24
clored-blocks

RustamKhananov:


src/style.css
```diff
@@ -0,0 +1,19 @@
+.red {
+  background-color: red;
+}
+
+.green {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.blue {
+  background-color: blue;
+}
+
+.noMargin {
+  margin: 0;
+}
+
+.high100 {
```
earthhuman41: i think it's bad idea to make this class, better to create common class for all divs and get them height 100px

src/index.html
```diff
@@ -8,7 +8,9 @@
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
-<body>
-  <h1>Colored blocks</h1>
+<body class="noMargin">
```
earthhuman41: no needed class here, if you want to reset default styles, just make 
```
body {
  margin: 0;
}
```

src/index.html
```diff
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <p class="upBlock noMargin"></p>
```
earthhuman41: why p instead of div?

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  background-color: green;
+}
+
+.bottomBlock {
+  background-color: blue;
+}
+
+.noMargin {
+  margin: 0;
+}
```
earthhuman41: you dont need this class

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+p {
```

src/index.html
```diff
@@ -1,14 +1,19 @@
...
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
+
 <body>
-  <h1>Colored blocks</h1>
+    <div class="upBlock"></div>
```
earthhuman41: Follow the indentation rules. Add an additional space of 2 spaces for nested tags.
https://mate-academy.github.io/style-guides/htmlcss.html#indentation

src/style.css
```diff
@@ -0,0 +1,20 @@
+
+body {
+  margin: 0;
+}
+
+div {
```

RustamKhananov: "earthhuman41 4 hours ago
i think it's bad idea to make this class, better to create common class for all divs and get them height 100px" 
Не понимаю, создать новый класс или прописать стиль для div?

earthhuman41: create new class, dont style tags

src/style.css
```diff
@@ -0,0 +1,20 @@
+
+body {
+  margin: 0;
+}
+
+.high100 {
```
earthhuman41: better .box or .block

PR #314, opened 2020-02-24
sol

vsm2k17:
https://vsm2k17.github.io/layout_colored-blocks/
https://vsm2k17.github.io/layout_colored-blocks/report/html_report/


src/index.html
```diff
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class = "r-block"></div>
```
vkryvytskyy:
```suggestion
  <div class="r-block"></div>
```
Useless spaces

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  width: auto;
+  height: 100px;
+  background-color: green;
+}
+
+.b-block {
+  width: auto;
+  height: 100px;
+  background-color: blue;
+}
```
vkryvytskyy: The name of the element class should make a sense(r-block, imho, doesn't). It should describe what the role of the element is on this page or describe its state. In the case with the name _red_, _green_, maybe in the future, the color of your block will be changed, so you have to rename all classes on the page.

PR #316, opened 2020-02-24
complete

OlEzhA0:
- [DEMO LINK](https://OlEzhA0.github.io/layout_colored-blocks/) </br>
    - [TEST REPORT LINK](https://OlEzhA0.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.first {
+  background-color: red;
+}
+
+.second {
+  background-color: green;
+}
+
+.last {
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
Ссылки должны быть кликабельными

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
```

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+.second {
+  background-color: green;
+  height: 100px;
+}
+
+.last {
+  background-color: blue;
+  height: 100px;
+}
```

PR #317, opened 2020-02-24
color was added to div

Nastia-rosy:
- [DEMO LINK](https://Nastia-rosy.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Nastia-rosy.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+.red {
+  background: red;
+}
+
+.green {
+  background: green;
+}
+
+.blue {
+  background: blue;
+}
+
+div {
+  height: 100px;
+}
+
+body {
+  margin: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background: green;
+}
+
+.blue {
+  background: blue;
+}
+
+div {
+  height: 100px;
+}
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background: red;
+}
+
+.green {
+  background: green;
+}
+
+.blue {
+  background: blue;
+}
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background: red;
+}
+
+#element_2 {
+  background: green;
+}
+
+#element_3 {
+  background: blue;
+}
```

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+.element_2 {
+  background: green;
+  height: 100px;
+}
+
+.element_3 {
+  background: blue;
+  height: 100px;
+}
```

PR #318, opened 2020-02-24
Develop

AlenaLoik:


vkryvytskyy:
Добавь в description пулл реквеста ссылки на DEMO и TEST RESULT

AlenaLoik:
Так правильно?

src/style.css
```diff
@@ -0,0 +1,25 @@
+
+body {
+  box-sizing: border-box;
+  width: 100%;
+  margin: 0;
+  padding: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,25 @@
+
+body {
+  box-sizing: border-box;
```
earthhuman41: it will apply only to body and this is redundant here

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  padding: 0;
+}
+
+div {
+  margin: 0;
+  padding: 0;
+  height: 100px;
+}
+
+.red {
```
earthhuman41: block--red better name

src/style.css
```diff
@@ -0,0 +1,21 @@
+
+body {
+  width: 100%;
+  margin: 0;
+  padding: 0;
```
earthhuman41: body hasn't default paddings

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.block--red {
+  background-color: red;
+  height: 100px;
+}
+
+.block--green {
+  background-color: green;
+  height: 100px;
```

PR #319, opened 2020-02-24
Solution colored-blocks

Nvaho:


src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  margin: -8px;
+  height: 5px;
+}
+
+#cell3 {
+  background-color: blue;
+  padding: 5%;
+  margin: -8px;
+  height: 5px;
+}
```

src/style.css
```diff
@@ -1,20 +1,17 @@
-#cell1 {
-  background-color: red;
+.Iblock {
   padding: 5%;
   margin: -8px;
   height: 5px;
 }
 
-#cell2 {
+.BlockOne {
```
vkryvytskyy: Названия классов пиши через kebab-case

Nvaho: я-учту-это-)

PR #320, opened 2020-02-24
Develop

IgorKulish:
- [DEMO LINK](https://IgorKulish.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://IgorKulish.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  background: green;
+}
+
+.blue {
+  background: blue;
+}
+
+body {
+  margin: 0;
+}
```
earthhuman41: style in the order in which the go elements in html

src/style.css
```diff
@@ -0,0 +1,20 @@
+div {
+  height: 100px;
+  width: auto;
+}
+
+.red {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+div {
+  height: 100px;
+  width: auto;
```
earthhuman41: Block elements has 100% of the width of the parent container. Set width auto here is redundant

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.block1,
+.block2,
+.block3 {
```
earthhuman41: create one common class

PR #321, opened 2020-02-24
message about changes

sanyanemesh:
- [DEMO LINK](https://sanyanemesh.github.io/layout_colored-blocks/)
    - [TEST REPORT LINK](https://sanyanemesh.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+#red-block {
+  height: 100px;
+  background-color: red;
+}
+
+#green-block {
+  height: 100px;
+  background-color: green;
+}
+
+#blue-block {
+  height: 100px;
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+#green-block {
+  height: 100px;
+  background-color: green;
+}
+
+#blue-block {
+  height: 100px;
+  background-color: blue;
+}
```

vkryvytskyy: The theme of the site may be changed and binding class names to colors is bad practice. You can think of these blocks as content, wrapper, and inner wrapper. Example: outer, inner, wrapper, content, etc.

vkryvytskyy: IDs are used for JS, try not to style the id if you can avoid it. This makes impossible to reuse the code. Stylize classes. Style guide

PR #322, opened 2020-02-24
solution

veli4kodima:


vkryvytskyy:
Ссылки на DEMO и TEST RESULT

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.square {
+  width: 100%;
+  height: 100px;
+
```
vkryvytskyy: nit(not important thing): remove empty line

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
```

PR #323, opened 2020-02-24
fix all errors

ShapovalDenys:
https://ShapovalDenys.github.io/layout_colored-blocks/

https://ShapovalDenys.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.red_rectangle {
+  background-color: red;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+.green_rectangle {
+  background-color: green;
+  height: 100px;
+}
+
+.blue_rectangle {
+  background-color: blue;
+  height: 100px;
+}
```

PR #324, opened 2020-02-24
index.html & style.css was updated

panasiuk-yuriy:
[DEMO LINK](https://panasiuk-yuriy.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://panasiuk-yuriy.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+#red {
+  background-color: red;
+  height: 100px;
+  width: 100%;
+}
+
+#green {
+  background-color: green;
+  height: 100px;
+  width: 100%;
+}
+
+#blue {
+  background-color: blue;
+  height: 100px;
+  width: 100%;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: green;
+  height: 100px;
+  width: 100%;
+}
+
+#blue {
+  background-color: blue;
+  height: 100px;
+  width: 100%;
+}
```

vkryvytskyy: IDs are used for JS, try not to style the id if you can avoid it. This makes impossible to reuse the code. Stylize classes. [Style guide](https://mate-academy.github.io/style-guides/htmlcss.html#id-selectors)

vkryvytskyy: The name of the element class should make a sense. It should describe what the role of the element is on this page or describe its state. In the case with the name _red_, _green_, maybe in the future, the color of your block will be changed, so you have to rename all classes on the page.

panasiuk-yuriy:
updated

src/style.css
```diff
@@ -2,20 +2,22 @@ body {
...
-#green {
+.green {
   background-color: green;
-  height: 100px;
-  width: 100%;
+
 }
 
-#blue {
+.blue {
```

PR #325, opened 2020-02-24
solution

ghost:
- [DEMO LINK](https://vitaliikorol.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://vitaliikorol.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  width: 100%;
+  height: 100px;
+}
+
+.first-block {
```
earthhuman41: bock--red will be better name

8hokage (changes requested):

earthhuman41:
Why are you ignoring this comment
Screen Shot 2020-02-24 at 3 40 11 PM

vitaliikorol:
div width removed, sorry for misunderstanding

8hokage (approved):

PR #326, opened 2020-02-24
message about changes

ilya-maker:
- [DEMO LINK](https://ilya-maker.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://ilya-maker.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+.blocks {
+  /*display: block;*/
+  /*width: 100%;*/
+}
+
+#block_1 {
+  content: " ";
+  height: 100px;
+  background-color: red;
+}
+
+#block_2 {
+  content: " ";
+  height: 100px;
+  background-color: green;
+}
+
+#block_3 {
+  content: " ";
+  height: 100px;
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  content: " ";
+  height: 100px;
+  background-color: green;
+}
+
+#block_3 {
+  content: " ";
+  height: 100px;
+  background-color: blue;
+}
```
vkryvytskyy: IDs are used for JS, try not to style the id if you can avoid it. This makes impossible to reuse the code. Stylize classes.  [style guide](https://mate-academy.github.io/style-guides/htmlcss.html#id-selectors)

ilya-maker:
Fixed

src/style.css
```diff
@@ -1,23 +1,17 @@
-.blocks {
-  /*display: block;*/
-  /*width: 100%;*/
-}
-
-#block_1 {
+.block {
   content: " ";
```
vkryvytskyy: this is redundant
image

PR #330, opened 2020-02-24
my solution

ZhupanovOlexii:
https://ZhupanovOlexii.github.io/layout_colored-blocks/
https://ZhupanovOlexii.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "files.autoSave": "afterDelay",
+    "files.autoSaveDelay": 500
+}
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class ="red"></div>
+  <div class ="green"></div>
+  <div class ="blue"></div>
 </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  background-color: red;
+  width: 100%;
+  height: 100px;
+}
+
+.green {
+  background-color: green;
+  width: 100%;
+  height: 100px;
+}
+
+.blue {
+  background-color: blue;
+  width: 100%;
+  height: 100px;
+}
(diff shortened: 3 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
vkryvytskyy: such files should not appear in the PR

src/index.html
```diff
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class ="red"></div>
```
vkryvytskyy:
```suggestion
  <div class="red"></div>
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: green;
+  width: 100%;
+  height: 100px;
+}
+
+.blue {
+  background-color: blue;
+  width: 100%;
+  height: 100px;
+}
```

src/index.html
```diff
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class ="block_1"></div>
```
vkryvytskyy:
```suggestion
  <div class="block_1"></div>
```
remove such spaces

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+.block_1 {
+  background-color: red;
+  width: 100%;
+  height: 100px;
+}
+
+.block_2 {
+  background-color: green;
+  width: 100%;
+  height: 100px;
```
vkryvytskyy: this pair of rules appears in three classes
If we have 5 blocks and these blocks have common characteristics, such as height, width, color, fonts, etc., then we should move these characteristics in a separate class and use it for style these blocks. DRY - Don't Repeat Yourself.

ZhupanovOlexii: Дякую, нам про це сьогодні розповіли. Зараз підправлю.

ZhupanovOlexii: Я поправив код, додав всім дівам ще один клас.

.vscode/settings.json
ZhupanovOlexii: я тут не розумію як і де їх прибрати

PR #331, opened 2020-02-24
Develop

OlenaZubashenko:
colored blocks

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.all {
+  display: block;
+  height: 100px;
+}
+
+.one {
+  background-color: red;
+}
+
+.two {
+  background-color: green;
+}
+
+.three {
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
Добавь в description ссылки на DEMO и TEST RESULT

OlenaZubashenko:
DEMO LINK https://olenazubashenko.github.io/layout_colored-blocks/
TEST REPORT LINK https://olenazubashenko.github.io/layout_colored-blocks/report/html_report/

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  background-color: red;
+}
+
+.two {
+  background-color: green;
+}
+
+.three {
+  background-color: blue;
+}
```

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.all {
+  display: block;
```
vkryvytskyy: this is default value for all div's, remove it

PR #333, opened 2020-02-24
my changes

cutiekate:
- [DEMO LINK](https://cutiekate.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://cutiekate.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.red {
+  height: 100px;
+  background-color: red;
+}
+
+.green {
+  height: 100px;
+  background-color: green;
+}
+
+.blue {
+  height: 100px;
+  background-color: blue;
+}
--- a/readme.md
+++ b/readme.md
@@ -1,13 +1,13 @@
 # Colored blocks
 1. Replace `<your_account>` with your Github username in the links
-    - [DEMO LINK](https://<your_account>.github.io/layout_colored-blocks/)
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  margin: 0;
+  padding: 0;
```
earthhuman41: most elements haven't default margins and padding, especially in this task you have just divs, so you don't need to reset it

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.red {
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.red {
```
earthhuman41: block--red is better name

src/style.css
cutiekate: fixed

src/style.css
cutiekate: fixed

src/style.css
cutiekate: fixed

src/style.css
cutiekate: fixed

src/style.css
```diff
@@ -0,0 +1,19 @@
+div {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+.size {
```
earthhuman41: its not size, you make styles for block so call it appropriate

PR #336, opened 2020-02-24
solution

TetianaZakharova:
[DEMO LINK](https://TetianaZakharova.github.io/layout_colored-blocks/)
    [TEST REPORT LINK](https://TetianaZakharova.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


TetianaZakharova:
[DEMO LINK](TetianaZakharova.github.io/layout_colored-blocks)
[TEST REPORT LINK](https://TetianaZakharova.github.io/layout_colored-blocks/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```

PR #337, opened 2020-02-24
'solution2'

FaychakA:


FaychakA:
- [DEMO LINK](https://FaychakA.github.io/layout_colored-blocks/)
    - [TEST REPORT LINK](https://FaychakA.github.io/layout_colored-blocks/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+  width: 100%;
```

8hokage (approved):

src/style.css
```diff
@@ -0,0 +1,19 @@
+.allMargin {
+  margin: 0;
+}
+
+.box_height {
```
earthhuman41: better just box

8hokage (approved):

src/style.css
FaychakA: ok

PR #338, opened 2020-02-24
solution2

AlexMachula:


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+}
+
+.block1 {
```
earthhuman41: better call it block--red

AlexMachula:
> - [DEMO LINK](https://AlexMachula.github.io/layout_colored-blocks/)
> - [TEST REPORT LINK](https://AlexMachula.github.io/layout_colored-blocks/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: redudant;
```
earthhuman41: you don't need this

earthhuman41: and there are no value redundant of width

PR #339, opened 2020-02-24
solution

OliinykKostya:


vkryvytskyy:
Ссылки на DEMO и TEST RESULT

OliinykKostya:
- [DEMO LINK](https://OliinykKostya.github.io/layout_colored-blocks/)
  - [TEST REPORT LINK](https://OliinykKostya.github.io/layout_colored-blocks/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+}
+
+.blue {
+  background-color: blue;
+}
+
+.green {
+  background-color: green;
+}
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px ;
+}
```

src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.firstBlock {
+  background-color: red;
+  height: 100px ;
+}
+
+.secondBlock {
```
vkryvytskyy: name classes using kebab-case

PR #341, opened 2020-02-24
2nd task to be done

Vlad-Viddi:
i also hope

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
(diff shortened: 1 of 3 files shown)


vkryvytskyy:
Ссылки на DEMO и TEST RESULT

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.colored-blocks__green {
+  background-color: green;
+  height: 100px;
+}
+
+.colored-blocks__blue {
+  background-color: rgb(0, 0,255);
+  height: 100px;
+}
```

PR #342, opened 2020-02-24
colored blocks

omelchenkom:


vkryvytskyy:
Ссылки на DEMO и TEST RESULT

omelchenkom:
а что с ссылками?

vkryvytskyy:
> а что с ссылками?

Их нет

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+.green {
+  height: 100px;
+  background-color: green;
+}
+
+.blue {
+  height: 100px;
+  background-color: blue;
+}
```

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: green;
+}
+
+.block3 {
+  background-color: blue;
+}
+
+div {
+  height: 100px;
+}
```

PR #343, opened 2020-02-24
Develop

YaroslavMakarov:


src/style.css
```diff
@@ -0,0 +1,21 @@
+.blue {
+  height: 100px;
+  background-color: blue;
+  width: 100%;
+}
+
+.green {
+  height: 100px;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+
+.red {
+  height: 100px;
+  background-color: red;
+  width: 100%;
+}
+
+body {
+  margin: 0;
+}
```
earthhuman41: style elems in order which they are in html in

src/style.css
```diff
@@ -0,0 +1,21 @@
+.blue {
+  height: 100px;
+  background-color: blue;
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.green {
+  height: 100px;
+  background-color: green;
+  width: auto;
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.red {
+  height: 100px;
+  background-color: red;
+  width: 100%;
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,19 @@
+.body {
+  margin: 0;
+}
+
+.div {
```

PR #344, opened 2020-02-24
message about changes

nchuhrina:


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  padding: 0;
+  margin: 0;
```
earthhuman41: You dont need to reset this in all elements

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+div {
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+}
+
+#red {
```
earthhuman41: class will be better

src/style.css
nchuhrina: changed reset only div

src/style.css
nchuhrina: now use class

PR #345, opened 2020-02-24
Solution

kochatkov:
- [DEMO LINK](https://kochatkov.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://kochatkov.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: 100;
+}
+
+.red {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.red {
```
earthhuman41: better call it block--red

PR #346, opened 2020-02-24
block solution

VakarchukDV:
- [DEMO LINK](https://VakarchukDV.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://VakarchukDV.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```

PR #347, opened 2020-02-24
solution

andriiFront:
- [DEMO LINK](https://andriiFront.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://andriiFront.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,23 @@
+body {
+  margin: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,23 @@
+body {
+  margin: 0;
+}
+
+h1 {
```

PR #353, opened 2020-02-25
Solution

vitaliyanohin:
[DEMO LINK](https://vitaliyanohin.github.io/layout_colored-blocks/)
 [TEST REPORT LINK](https://vitaliyanohin.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.box {
+  display: grid;
+  width: 100%;
+  height: 100px;
+}
+
+.item1 {
+  background: red;
+}
+
+.item2 {
+  background: green;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.box {
+  display: grid;
```
vkryvytskyy: why you use `display: grid`? it's redundant here

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.box {
+  width: 100%;
```
vkryvytskyy: width 100% is default value for all div's

PR #356, opened 2020-02-25
Colored blocks

volodymyrtaras:
https://volodymyrtaras.github.io/layout_colored-blocks/
https://volodymyrtaras.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+#first,
+#second,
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  background-color: red;
+}
+
+#second {
+  background-color: green;
+}
+
+#third {
+  background-color: blue;
+}
```

PR #362, opened 2020-04-17
Solution

aleksey-10:
- [DEMO LINK](https://aleksey-10.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://aleksey-10.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+* {
+  margin: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+* {
```
YuriiHoliuk, line 17: Do not use the asterisk selector(`*`) without necessary. It could be used only for `box-sizing: border-box;`

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
```
YuriiHoliuk, line 15: It is ok for this task(leave it as is), but usually using style in classname is a bad idea

PR #368, opened 2020-04-21
Solution

Mishtal-Andrii:
- [DEMO LINK](https://Mishtal-Andrii.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://Mishtal-Andrii.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,19 @@
 <!doctype html>
 <html lang="en">
+
 <head>
   <meta charset="UTF-8">
   <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
+
 <body>
-  <h1>Colored blocks</h1>
+  <div class="block top"></div>
+  <div class="block middle"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
```
YuriiHoliuk, line 1: Try to avoid selector *. Yes in these tasks it doesn't influence performance, but in bigger projects, it can load browser

src/index.html
```diff
@@ -1,14 +1,19 @@
...
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
+
 <body>
-  <h1>Colored blocks</h1>
+  <div class="block top"></div>
+  <div class="block middle"></div>
+  <div class="block bottom"></div>
```
YuriiHoliuk, line 16: Great classnames👍

src/style.css
Mishtal-Andrii: Ok

src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: 100%;
```
irenhh, line 6: it's 100% wide by default

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  height: 100px;
+  margin: 0;
```
irenhh, line 8: remove margin resetting here

src/style.css
Mishtal-Andrii: ok

src/style.css
Mishtal-Andrii: ok

PR #369, opened 2020-04-21
Solution

V2473:
- [DEMO LINK](https://V2473.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://V2473.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0 !important;
+  padding: 0 !important;
```
irenhh, line 3: don't use `!important`, it is difficult to override

irenhh: body doesn't have default paddings

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+div {
```
irenhh, line 6: use general class name

irenhh (changes requested):

irenhh (approved):

PR #370, opened 2020-04-21
Solution

Kussmich:
- [DEMO LINK](https://kussmich.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://kussmich.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: 100%;
+  height: 100px;
+}
+
+.block:first-of-type {
+  background-color: red;
+}
+
+.block:nth-of-type(2) {
+  background-color: green;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.block {
+  width: 100%;
```
irenhh, line 6: it's a default value for block element

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+.block:first-of-type {
+  background-color: red;
+}
+
+.block:nth-of-type(2) {
+  background-color: green;
+}
+
+.block:last-of-type {
```
irenhh, line 18: better to add class names

PR #371, opened 2020-04-21
Solution

germandolia:
[DEMO LINK](https://germandolia.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://germandolia.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+.nomargin {
+  margin: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+.nomargin {
```
YuriiHoliuk, line 17: it is ok to style the body, but the class name is a great idea too. Better to choose the name `margin-0`

src/style.css
```diff
@@ -0,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+.red {
```
YuriiHoliuk, line 5: it is ok in this task but usually using a style in class name is a bad idea

PR #372, opened 2020-04-21
colored block Ivanchenko

tanyamoore:
https://tanyamoore.github.io/layout_colored-blocks/
https://tanyamoore.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.box {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+
+.box {
+  height: 100px;
+  width: 100%;
```
irenhh, line 8: it's a default behavior

src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
+  margin: 0;
+}
```
irenhh, line 3: you don't need to reset it everywhere

irenhh: it makes your code slower if you do extra actions

PR #373, opened 2020-04-21
Colored blocks

LimonovInsp:
- [DEMO LINK](https://LimonovInsp.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://LimonovInsp.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
+  padding: 0;
+  margin: 0;
+}
+
+.firstBlock {
+  background-color: red;
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk, line 9: [DRY](https://www.google.com/search?q=DRY&oq=DRY&aqs=chrome..69i57j0l7.1684j0j7&sourceid=chrome&ie=UTF-8) - do not repeat code

src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
```

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.firstBlock {
```
YuriiHoliuk, line 6: classnames should be in `snake_case` or in `kebab-case`, not in `camelCase`

PR #374, opened 2020-04-21
Colored blocks added

SerhiiiOselskyi:
- [DEMO LINK](https://bran4u.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://bran4u.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+
+.red-box {
+  background-color: red;
+  width: auto;
```
irenhh, line 8: it is unnecessary, width is 100% by default

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  height: 100px;
```
irenhh, line 9: create a common class, which has this style and use it everywhere you need this style

irenhh (changes requested):

src/index.html
```diff
@@ -8,7 +8,9 @@
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
-<body>
-  <h1>Colored blocks</h1>
+<body style="margin: 0; padding: 0;">
+  <div style="background-color: red; height:100px;"></div>
+  <div style="background-color: green; height:100px;"></div>
+  <div style="background-color: blue; height:100px;"></div>
```
irenhh, line 14: why did you decide to use inline styles?

use a separate file with styles

irenhh (changes requested):
and don't forget about the previos comments

src/index.html
bran4u: because common class will got only height:100px, and i should set background color for each.

src/index.html
```diff
@@ -8,7 +8,9 @@
   <title>Colored blocks</title>
   <link rel="stylesheet" href="style.css">
 </head>
-<body>
-  <h1>Colored blocks</h1>
+<body style="margin: 0; padding: 0;">
```
irenhh, line 11: all styles in the separate file

src/style.css
```diff
@@ -0,0 +1,14 @@
...
+
+.red-background{
+  background-color: red;
+}
+.green-background{
+  background-color: green;
+}
+.blue-background{
+  background-color: blue;
+}
```
irenhh, line 14: check indentations and code formatting, run linter, add new line in the end of the file

irenhh (changes requested):

irenhh (changes requested):

irenhh (approved):

PR #375, opened 2020-04-21
Task

VikRadel:
[Demo Link](https://vikradel.github.io/layout_colored-blocks/)
[Demo Link Test](https://vikradel.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
+  width: 100%;
```
irenhh, line 6: it's a default value for block element

VikRadel:
Updated, should be fine now

PR #376, opened 2020-04-21
Solution

NazarDovhanchuk:
- [DEMO LINK](https//nazardovhanchuk.github.io/layout_colored-blocks/)
    - [TEST REPORT LINK](https://nazardovhanchuk.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  display: block;
+  height: 100px;
+  background-color: red;
+}
+
+.green {
+  display: block;
+  height: 100px;
+  background-color: green;
+}
+
+.blue {
+  display: block;
+  height: 100px;
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+body {
+  margin: 0;
+}
+
+.red {
+  display: block;
```
irenhh, line 6: use div to avoid this

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  height: 100px;
```
irenhh, line 7: use general class name

src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.box_1 {
```
irenhh, line 5: this is not a good idea to use numbers to identify the element, use names that describe the content or its behavior

PR #380, opened 2020-04-21
Colored blocks denkondratiev

denkondratiev:
- [DEMO LINK](https://denkondratiev.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://denkondratiev.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class="red"></div>
+  <div class="green"></div>
+  <div class="blue"></div>
 </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
YuriiHoliuk, line 5: Avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins. Setting width and height for all DIV tags is bad practice

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  height: 100px;
+}
+
+.red {
```
YuriiHoliuk, line 9: it is ok in this task but usually using a style in class name is a bad idea

src/index.html
```diff
@@ -9,6 +9,8 @@
   <link rel="stylesheet" href="style.css">
 </head>
 <body>
-  <h1>Colored blocks</h1>
+  <div class="r h"></div>
+  <div class="g h"></div>
+  <div class="b h"></div>
```
irenhh, line 14: how do you think other developers know what you mean by those names?

you should give clear and descriptive names

PR #381, opened 2020-04-21
Colored blocks solution

tyooma:
- [DEMO LINK](https://tyooma.github.io/layout_colored-blocks/) <br>
- [TEST REPORT LINK](https://tyooma.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.red {
```
YuriiHoliuk, line 9: it is ok in this task but usually using a style in class name is a bad idea

src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
```

PR #383, opened 2020-04-21
 layout_colored-blocks by Nickyforov

5lum6er:
- [DEMO LINK](https://5lum6er.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://5lum6er.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,22 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
+  width: 100%;
+}
+
+.green {
+  background-color: green;
+  height: 100px;
+  width: 100%;
+}
+
+.blue {
+  background-color: blue;
+  height: 100px;
+  width: 100%;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.red {
+  background-color: red;
+  height: 100px;
+  width: 100%;
```
YuriiHoliuk, line 9: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - do not repeat code

src/style.css
```diff
@@ -0,0 +1,22 @@
+* {
```

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.red {
```
YuriiHoliuk, line 6: it is ok in this task but usually using a style in class name is a bad idea

src/style.css
```diff
@@ -0,0 +1,26 @@
+body,
+.first,
+.second,
+.third {
+  padding: 0;
+  margin: 0;
+}
```
irenhh, line 7: the only one that needs resetting margins is `body`, don't apply this to everything;
paddings resetting is not necessary here at all

src/style.css
irenhh: create one general class name for repeated styles

PR #384, opened 2020-04-21
solution

DimaDP:
**- [DEMO LINK](https://dimadp.github.io/layout_colored-blocks/)**
  **- [TEST REPORT LINK](https://dimadp.github.io/layout_colored-blocks/report/html_report/)**

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+* {
+  margin: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+.red {
```
YuriiHoliuk, line 5: it is ok in this task but usually using a style in class name is a bad idea

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+* {
```
YuriiHoliuk, line 17: Try to avoid selector *. Yes in these tasks it doesn't influence performance, but in bigger projects, it can load browser

PR #386, opened 2020-04-21
Develop

lenatao15:
- [DEMO LINK](https://lenatao15.github.io/layout_colored-blocks/)
   - [TEST REPORT LINK](https://lenatao15.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+div {
```
YuriiHoliuk, line 5: Try to avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins. Setting width and height for all DIV tags is bad practice

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  height: 100px;
+  width: 100%;
+}
+
+.red {
```
YuriiHoliuk, line 10: it is ok in this task but usually using a style in class name is a bad idea

src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.flag {
+  height: 100px;
+  width: 100%;
```
irenhh, line 7: block element has 100% width by default

PR #387, opened 2020-04-21
add color blocks

wa7er14:
[DEMO LINK](https://wa7er14.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://wa7er14.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.green {
+  background-color: green;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+div {
```
irenhh, line 6: use class name for this, you may have more divs and other ones will have different appearance, this way you will need to override your styles

PR #388, opened 2020-04-21
3 colorful blocks

hryshko-denys:
https://github.com/hryshko-denys/layout_colored-blocks
https://hryshko-denys.github.io/layout_colored-blocks/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: 0;
+  padding: 0;
```
irenhh, line 3: body doesn't have default paddings

PR #389, opened 2020-04-21
blocks

YanG-6668:
- [DEMO LINK](https://YanG-6668.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://YanG-6668.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,23 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  height: 100px;
+  width: 100%;
+  padding: 0;
+  margin: 0;
+}
+
+.block1 {
+  background-color: red;
+}
+
+.block2 {
+  background-color: green;
+}
+
+.block3 {
+  background-color: blue;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,23 @@
+body {
+  margin: 0;
+  padding: 0;
```
irenhh, line 3: body doesn't have default paddings

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.block {
+  height: 100px;
+  width: 100%;
```
irenhh, line 8: width: 100% is a default value for block elements

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+  padding: 0;
+  margin: 0;
```
irenhh, line 10: there is no need to reset paddings and margins for div

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.block1 {
```
irenhh, line 13: don't use numbers to identify the element, use names that describe the content or the behavior of the element

PR #390, opened 2020-04-21
Solution

YarosKr:
- [DEMO LINK](https://YarosKr.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://YarosKr.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,22 @@
+
+.block1 {
+  width: 100%;
```
irenhh, line 3: block element has such width by default

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  height: 100px;
```
irenhh, line 4: create a general class name for this style as it's repeated several times

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  background: green;
+}
+
+.block3 {
+  width: 100%;
+  height: 100px;
+  background: blue;
+}
+
+body {
```
irenhh, line 20: styles for body are defined on the top of the document

src/style.css
```diff
@@ -0,0 +1,22 @@
+
+.block1 {
+  width: 100%;
+  height: 100px;
+  background: red;
+}
+
+.block2 {
```
irenhh, line 8: it is not a good idea to identify elements using numbers, give the name that describes the content or behavior of the element

PR #405, opened 2020-06-27
solution

nanccyy:
- [DEMO LINK](https://nanccyy.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://nanccyy.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,18 @@
+body {
+  margin: 0;
+}
+
+.red-block {
+  background-color: red;
+  height: 100px;
```
Denchepornyuk, line 7: you use `height: 100px;` three times
better to create class `block` and add this class to each `div` on the page

Example:
```
block {
  height: 100px;
}

.red-block {
  background-color: red;
}

...
```

src/index.html
```diff
@@ -9,6 +9,8 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <div class="red-block"></div>
```
Denchepornyuk, line 12: add class `block` to each div

PR #406, opened 2020-06-27
Solution

Mikle25:
[DEMO LINK](https://Mikle25.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://Mikle25.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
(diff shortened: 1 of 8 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+}
+
+div {
```
irenhh, line 5: usually you will need a general class for elements with the same styles,
styling a tag is not the best choice since you may have other div tags that don't need this style

irenhh (changes requested):

irenhh (approved):

PR #408, opened 2020-06-27
Solution

mpone:
- [DEMO LINK](https://mpone.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://mpone.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+* {
+  margin: 0;
+}
+
+.block {
+  height: 100px;
+}
+
+.color-red {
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
```
Denchepornyuk, line 1: `*` selector is used to add this styles to `EACH` element of the file, you don't usually need this and this is too heavy for the big page, that's why this is better to add this style to those elements you really need

PR #416, opened 2020-06-27
Solution

zkhomer:
- [DEMO LINK](https://zkhomer.github.io/layout_colored-blocks/)
 - [TEST REPORT LINK](https://zkhomer.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,8 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <div class="redBlock"></div>
+    <div class="greenBlock" ></div>
+    <div class="blueBlock"></div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+
+* {
+  margin: 0;
+}
+
+.redBlock {
+  background-color: red;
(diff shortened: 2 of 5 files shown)


src/index.html
```diff
@@ -9,6 +9,8 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <div class="redBlock"></div>
```
irenhh, line 12: class names in html ususally use `snake_case` or `kebab-case`

src/style.css
```diff
@@ -0,0 +1,19 @@
+
+* {
```
irenhh, line 2: `*` selector is used to add this styles to EACH element of the file, you don't usually need this and this is too heavy for the big page, that's why this is better to add this style to those elements you really need

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.greenBlock {
+  background-color: green;
+  height: 100px;
+}
+
+.blueBlock {
+  background-color: blue;
+  height: 100px;
+}
```
irenhh, line 19: all them have similar property, which can be moved to one general class and this class can be added to all the elements that need this style

for example `block`

PR #418, opened 2020-06-28
Solution

1nsider21:
- [DEMO LINK](https://1nsider21.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://1nsider21.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,18 @@
+* {
+  margin: auto;
+}
```
irenhh, line 3: - `*` selector is used to add this styles to EACH element of the file, you don't usually need this and this is too heavy for the big page, that's why this is better to add this style to those elements you really need

src/style.css
```diff
@@ -0,0 +1,18 @@
...
+
+.green {
+  background-color: green;
+  height: 100px;
+}
+
+.blue {
+  background-color: blue;
+  height: 100px;
+}
```
irenhh, line 18: - all them have similar property, which can be moved to one general class and this class can be added to all the elements that need this style

irenhh: you can add one more class like `block` and add height style to it

PR #424, opened 2020-06-29
Develop

b0hdnan:
- [DEMO LINK](https://b0hdnan.github.io/layout_colored-blocks/)

-  [TEST REPORT LINK](https://b0hdnan.github.io/layout_colored-blocks/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,19 @@
+body {
+  margin: 0;
+  padding: 0;
+}
```
irenhh, line 4: `body` doesn't have default paddings, do you don't need to remove them

src/style.css
```diff
@@ -0,0 +1,19 @@
...
+
+.block2 {
+  background-color: green;
+  height: 100px;
+}
+
+.block3 {
+  background-color: blue;
+  height: 100px;
+}
```
irenhh, line 19: - all them have similar property, which can be moved to one general class and this class can be added to all the elements that need this style
- use one more class like `block` that will have similar styles

src/index.html
```diff
@@ -9,6 +9,8 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <div class="block1"></div>
+    <div class="block2"></div>
+    <div class="block3"></div>
```
irenhh, line 14: don't give names like `name1`, `name2`, etc.
use only meaningful names

PR #427, opened 2020-06-29
Solution

martakupka:
- [DEMO LINK](https://martakupka.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://martakupka.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,13 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <!-- <h1>Colored blocks</h1> -->
+
+    <div id="block-container">
+      <div id="block-1"></div>
+      <div id="block-2"></div>
+      <div id="block-3"></div>
+    </div>
+
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+body {
+  margin: 0;
+}
+
+#block-container {
+  display: flex;
+  flex-direction: column;
+}
+
+#block-1 {
+  background-color: red;
+  width: 100%;
+  height: 100px;
+}
+
+#block-2 {
+  background-color: green;
+  width: 100%;
+  height: 100px;
+}
+
+#block-3 {
+  background-color: blue;
+  width: 100%;
+  height: 100px;
+}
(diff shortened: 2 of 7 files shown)


src/index.html
```diff
@@ -9,6 +9,13 @@
     <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Colored blocks</h1>
+    <!-- <h1>Colored blocks</h1> -->
```
irenhh, line 12: remove this

src/index.html
```diff
@@ -9,6 +9,13 @@
...
+
+    <div id="block-container">
+      <div id="block-1"></div>
+      <div id="block-2"></div>
+      <div id="block-3"></div>
+    </div>
+
```
irenhh, line 19: - don't give names like `block1`, `block2`, etc.; use only meaningful names

src/style.css
```diff
@@ -0,0 +1,26 @@
+body {
+  margin: 0;
+}
+
+#block-container {
```
irenhh, line 5: don't style with ids, they have quite a high specificity so they are harder to override with other styles

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  display: flex;
```
irenhh, line 6: do you really need this style?

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  flex-direction: column;
```
irenhh, line 7: the same

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  background-color: green;
+  width: 100%;
+  height: 100px;
+}
+
+#block-3 {
+  background-color: blue;
+  width: 100%;
+  height: 100px;
+}
```
irenhh, line 26: - all them have similar property, which can be moved to one general class and this class can be added to all the elements that need this style

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+
+#block-container {
+  display: flex;
+  flex-direction: column;
+}
+
+#block-1 {
+  background-color: red;
+  width: 100%;
```
irenhh, line 12: - div is a block element, block elements are 100% wide

PR #451, opened 2020-06-30
colored blocks solution

asbh25:
- [DEMO LINK](https://asbh25.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://asbh25.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+* {
+  margin: 0;
+}
+
+div {
+  height: 100px;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
+  margin: 0;
+}
+
+div {
```
Denchepornyuk, line 5: не стилізуй теги

src/style.css
```diff
@@ -0,0 +1,19 @@
+* {
```
Denchepornyuk, line 1: не використовуй `*`

PR #524, opened 2020-11-02
Added background-color

SofiiaAsieieva:
[DEMO LINK](https://SofiiaAsieieva.github.io/layout_colored-blocks/)
[TEST REPORT LINK](https://SofiiaAsieieva.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,19 @@
+.block {
+  height: 100px;
+}
+
+body {
+  margin: 0%;
+}
+
+#one {
+  background-color: red;
+}
+
+#two {
+  background-color: green;
+}
+
+#three {
+  background-color: blue;
+}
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,19 @@
...
+  background-color: red;
+}
+
+#two {
+  background-color: green;
+}
+
+#three {
+  background-color: blue;
+}
```
irenhh, line 19: usually, you don't use id for styling since it has a higher specificity, use classes instead

irenhh (approved):

src/style.css
SofiiaAsieieva: Спасибо, буду знать😊

PR #545, opened 2020-11-08
three colored blocks was added

NikitaTudakov:
- [DEMO LINK](https://NikitaTudakov.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://NikitaTudakov.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: auto;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
+
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
+body {
+  margin: auto;
+}
+
+div {
+  height: 100px;
+  width: 100%;
```
vkryvytskyy, line 7: div by default has width 100%

vkryvytskyy (approved):

src/style.css
NikitaTudakov: Принято!

PR #549, opened 2020-11-09
three coloured blocks added

taniazt:
- [DEMO LINK](https://taniazt.github.io/layout_colored-blocks/)
- [TEST REPORT LINK](https://taniazt.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+.red {
+  background-color: red;
+}
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+div {
+  height: 100px;
+  width: 100%;
+}
+
+body {
+  margin: 0;
+}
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
...
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+div {
+  height: 100px;
+  width: 100%;
```
vkryvytskyy, line 15: this is default value for all div's

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+.green {
+  background-color: green;
+}
+
+.blue {
+  background-color: blue;
+}
+
+div {
```
vkryvytskyy, line 13: use classes for styling, don't style elements directly

vkryvytskyy (changes requested):

src/style.css
taniazt: fixed

src/style.css
taniazt: fixed

vkryvytskyy (approved):

PR #571, opened 2020-11-10
solution ('colored clocks')

5Mountains:
- [x] [DEMO LINK](https://5mountains.github.io/layout_colored-blocks/)
 - [x] [TEST REPORT LINK](https://5mountains.github.io/layout_colored-blocks/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  width: 100%;
+  height: 100px;
+}
+
+.block__red {
+  background: red;
+}
+
+.block__green {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,21 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.block {
+  width: 100%;
```
vkryvytskyy, line 7: this is default value for all div's

vkryvytskyy (approved):

src/style.css
5Mountains: Agree with you! My bad habit *(

